PR #7, opened 2019-01-24
Added /dev/input/mice driver for linux

kurtnalty:
1) Added mou_devmice.c driver for linux using /dev/input/mice
2) Added mou_devmouse section to src/drivers/Objects.rule
3) Added config.linux-fb
4) Added comma to line 70, src/drivers/vtswitch.c
5) Removed extra 1on line 158, src/drivers/kbd_tty.c

### DIFF
--- a/src/Configs/config.linux-fb
+++ b/src/Configs/config.linux-fb
@@ -1,0 +1,333 @@
+####################################################################
+# config - Microwindows and Nano-X configuration file
+#
+# Set target architecture using ARCH= from options in Arch.rules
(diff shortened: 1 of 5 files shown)


ghaerr:
Kevin,

Thanks for your patch. I’ve done some large changes lately
to src/drivers/Objects.rules and config formats, so I’ll modify your
patch on those files slightly and then apply it, while also adding your mou_devmice.c
driver, kbd_tty.c and vtswitch.c fixes.

Thanks!

Regards,

Greg

ghaerr:
Thanks Kurt, updated patch to latest source and did commit.

PR #9, opened 2019-01-29
Check for valid window before doing some actions

roozbehid:
These set of checks for valid window was required during showing of pop ups in Webassembly.
When popup window get destroyed, something would still try to send msg to them -which could be fault of my C# code- and will cause a crash!


ghaerr:
Added

PR #12, opened 2019-01-29
Make MWBUTTON defines compatible with windows

roozbehid:
These definitions were not compatible with what was used in Windows as they were passed back to my system they would cause different really hard to pin problems!


ghaerr:
Rejected. You have MWBUTTON_M defined as "10" which has two bits set: 8 and 2. Did you mean 0x10 or 010? This cannot be the value Windows uses.

PR #19, opened 2019-02-07
VisualStudio contrib and more Rooz stuff

roozbehid:
So I am redoing stuff based on your master. This is mostly minimum stuff required to make it compile/build with Visual Studio.
If merged, then I will send subsequent pull request for other stuff to make it actually usable for me. This is bare minimum for me to compile it.


ghaerr:
Thanks for the clean modifications :)

Are you setting -DWINEXTRA=1 and -DWIN32=0 (or not defining WIN32 at all)?

The WINEXTRA=1 define causes my windows.h to include winextra.h at the end, which should negate the need for your including it again in the contrib/winextra/stub*.c routines. Also, this define adds in your required changes to the Microwindows win32 internals. I don't understand the need for the #pragma once in each of the contrib/winextra/*.h files, these shouldn't be included twice?

I am recommending against using WIN32 defined at the moment, as it is used in several places that need cleaning up and don't want dragged into your project (like nanox/srvmain.c GsSelect() where you've had to add <windows.h> and nx11/X11-local/X11/Xos.h commenting out an include).

If we just stick with using #if WINEXTRA and not WIN32 for now, it will be easy to track your changes internally. I'm going to deprecate the WIN32 use soon as it is internally confusing. Of course, we're already using #if  _MSC_VER for those cases where the change is based on the MSVC build environment and header files.

roozbehid:
So I know for sure I had -DWINEXTA -DWIN32 in all projects, I think it ends up =1. I will check that. But pragma once was necessary. It seems they were included twice. I'll check on these.

A bit of problem with WIN32 is that all projects by default when created in visual studio would inherit WIN32. So you have to go and manually remove them.

ghaerr:
> So I know for sure I had -DWINEXTA -DWIN32 in all projects, I think it ends up =1. I will check that. But pragma once was necessary. It seems they were included twice. I'll check on these.
> 
> 

Yes, please check that WINEXTRA is 1, it won’t work otherwise, as I added
all the internals you used in your tree using this define “#if WINEXTRA” for inclusion when =1.
> A bit of problem with WIN32 is that all projects by default when created in visual studio would inherit WIN32. So you have to go and manually remove them.
> 
> 

Don’t worry about WIN32 having to be 0, it’s not worth having to go against
the default setup for visual studio. At this point, it doesn’t matter what the setting
is, and I will be removing all WIN32 #if/#ifdef from other ancillary code shortly.

Take a look at src/docs/ROOZ.txt section 1, you’ll want to set the other -D defines I’ve recommended
since you’re not using a config file and Arch.rules/Makefile.rules which setup required
compilation defines.

Let me know what you think about the info in the other sections.

For the many other windows .h files you need to compile your project, you might
want to consider sticking them into another include directory in your project, rather
than src/contrib/winextra, unless they are part of the core Win32 .h files Microwindows
already handles.

PR #20, opened 2019-02-12
Added SDL support for android. (added proper config aswell)

nwrkbiz:
I successfully created an fltk android app using SDL.

SDL seems to be much faster and more stable than the allagro backend (subjective feeling) :)

Things that work better with SDL:
- App does not freeze when put into background 
- App does not freeze on rotate


nwrkbiz:
something like the zoom factor introduced to allegro would be awesome here aswell (see attached screenshot).
[screenshot_2019-02-12_204915](https://user-images.githubusercontent.com/6352567/52663912-10d2b900-2f08-11e9-965e-b0f20b0cc6f5.jpg)
)

ghaerr:
Daniel,

I'm going to accept this, but will need to bracket the {SDL,ALLEGRO}_{INCLUDE,LIB} with not-blank testing, otherwise this breaks the build for desktop SDL and ALLEGRO by introducing an empty "-I" that eats the next compiler argument for config files that don't have the setting. I will have time to do this shortly.

SDL is definitely more stable than Allegro on my OS X desktop, although Allegro's speed seems pretty good. There are mouse cursor bugs on Allegro desktop I can't seem to perfect.

I will try adding zoom on SDL which will help you. Allegro runs smaller on my retina displays so it was easy to implement, but SDL displays a larger size automatically on OSX.

nwrkbiz:
What is the default/right way to handle dependencies (allegro, sdl) for the desktop builds? I guess the libraries + headers need to be installed into the system path, am I right? I wanted to avoid this by specifying the includes in the config. Not sure if this is a neat approach at all.

The zoom thing was just an idea, because i really liked it with allegro :) There is no urgent need for me having this feature, i was just curious if fltk + microwindows + sdl works like it does with allegro and I am impressed, it does :D

Next thing I want to try is, to run my template project via a NOMMU linux on a nintendo ds console using your awesome microwindows architecture.

Thanks again for the effort you are putting into this project! :)

ghaerr:
There is no "right" way to handle the different includes for libraries, for the old-time fonts and images we have been using config file options, and using SDL2_INCLUDE/LIB works well. I updated Arch.rules to protect the build when these defines are empty, which is the case for the desktop builds where the libs are usually installed into the system path.

I added the zoom feature to SDL2 for you and defaulted it to 2.0 for Android :)

I'd like to see the NDS support, let me know how it works and go ahead and update Arch.rules in the ARCH=NDS section if you need to.

ghaerr:
I forgot to mention that with SDL2, when zoom is not 1.0, you will likely need to adjust your SCREEN_WIDTH and HEIGHT (by halving them) in order for things to look like Allegro. This is because SDL seems to automatically adjust the display (at least on OSX) to a larger size for zoom 1.0, where Allegro does not. Thus, the screen display size "doubling" trick I used in Allegro has SDL doubling the screen size while also zooming.

Let me know how it works for you.

nwrkbiz:
Works perfectly! hat to divide SCREEN_WIDTH/HEIGHT by two as suggested by you.

See the result:
screenshot_20190213-032137

PR #30, opened 2019-05-22
Adding AquilaOS support

mohamed-anwar:


ghaerr:
Clean work, thanks!
I've used the convention of #if <platform> rather than #ifdef <platform> internally. That means that your Arch.rules CFLAGS -DUNIX should probably be changed to -DUNIX=1 (not sure if all compilers set =1 when left blank). Also helps to do the same with AQUILA to avoid mixing ifdef/if's. FYI.

If you're going to be using Nuklear, you'll want to add mousewheel support for sure. Simply done by returning an additional signed wheel up/down value as the fourth byte. See drivers/mou_devmice.c or drivers/mou_sdl2.c for details. We currently ignore the value magnitude, just use the sign to set the MWBUTTON_SCROLLUP/DN flag, and the rest is handled internally. You'll also need to update the GetButtonInfo with the two new flags.

PR #43, opened 2020-08-28
Portrait updates

djipi:
Added portrait compilation directives.


ghaerr:
This looks good, but I would like to see a lot less \#ifdefs. As you have seen, except for memory constrained systems, I much prefer staying away from adding \#ifdefs to the core code. (Yes, there are already too many!)

That said, and given that Microwindows needs to run on memory-constrained systems, adding a configurable portrait mode feature is still a good idea, as there is considerable code added in each of the sub drivers which are not needed for some systems.

I would like to suggest the following edits for all sub drivers, please:

1) Instead of using three \#ifs on each of the static drivers, use a single \#if around all three of them.

2) In the final PSUBDRIVER function array in each driver, how about something like the following, for better readability:
```
PSUBDRIVER fblinear1[4] = {
      &fblinear1_none
#if MW_FEATURE_PORTRAIT	
 	, &fbportrait_left, &fbportrait_right, &fbportrait_down
#endif
```
Uninitialized data is always inited to 0, the the lack of explicit NULLs will not be a problem.

Finally, it looks like the .gitignore issues are already handled in your PR #42.

djipi:
I have updated the files in accordance for the #ifdef's.
I have notice the .gitignore repeat in the branch; it comes probably from the branching I'm doing.

ghaerr:
Thanks @djipi,

We're almost there... it looks like you might have forgot to update fblin8.c and fblin32.c.

djipi:
Sorry @ghaerr , should be fine now.

PR #44, opened 2020-08-29
Support & fix

djipi:
The potential fix for the fonts is linked to the function strncpy. A second look will be welcome.
"No null-character is implicitly appended at the end of destination if source is longer than num."


rofl0r:
rebased on fixed master as #45. can we move discussion over there ?

PR #45, opened 2020-08-29
Support & fix (re-upload of #44)

rofl0r:
1) Add BMP format compilation in case of HAVE_BMP_SUPPORT is set to N.
2) Add Visual Studio / Win32 support in the BMP file converter souce.
3) Fix in gen_getscreeninfo() to add a break within 'case MWIF_RGBA8888'.
4) Potential fix for a string manipulation in the fonts.

PR #49, opened 2020-09-03
Fixes for string operations

djipi:
1) Fixed a typo error in Readme.md
2) String manipulation fix previously discussed in #45.

Note: There are still some strncpy function used in the code. You may want to replace those by strzcpy.


ghaerr:
Thanks @djipi!

ghaerr:
> There are still some strncpy function used in the code. You may want to replace those by strzcpy.

Whoops, missed that... The font name copying functions are a little tricky, as Win32 doesn't expect them to be NUL terminated. I assume this last fix was the result of an error with a font name on your platform?

Can you point out a few examples that fail? I see there are others in various places.

djipi:
I do not have dedicated examples unfortunately; the problem occured during my early test of the port. Basically the string was damaged at my end, and got the error. So, I thought it was probably better to prevent overflow in case of a long and valid string.

Issue #55, opened 2020-10-16
Compile on Ubuntu for Emscripten


AdlIzzadin:
I cant compile MicroWindows on Ubuntu for Emscripten.

Linking mwapp.html ...
emcc  -O3 -I./nx11/X11-local -I/usr/include/freetype2 -I/usr/include/libpng16 -lfreetype -s USE_SDL=2 -s USE_ZLIB=1 -s USE_FREETYPE=1 -s USE_LIBPNG=1 -I/usr/include/freetype2 -I/usr/include/libpng16 -lfreetype  -Iinclude -I/usr/include/freetype2 -I/usr/include/libpng16 -lfreetype -DEMSCRIPTEN=1 -DMULTIAPP=1 -DHAVE_FREETYPE_2_SUPPORT=1	   -s WASM=0 -s TOTAL_MEMORY=67108864 -s ERROR_ON_UNDEFINED_SYMBOLS=0 -o bin/mwapp.html ./demos/mwin/mwapp.c ./demos/mwin/mwdemo.c ./demos/mwin/mwdemo2.c ./demos/mwin/mwmine.c ./images/demos/mwin/mwmine/mineflag.c ./images/demos/mwin/mwmine/mineface.c ./images/demos/mwin/mwmine/minefacelost.c ./images/demos/mwin/mwmine/minebomb.c ./images/demos/mwin/mwmine/minedone.c ./images/demos/mwin/mwmine/minehitfalse.c lib/libmwin.a
**wasm-ld: error: unknown file type: winmain.o**
[...]

PR #57, opened 2020-12-28
Update build for Emscripten v2.0.11

### LINKED ISSUE #37: Errors compiling with emscripten
It looks like the emscripten build might have some issues on Macos X 10.15.2. I'm using the latest emsdk and running the 'make-emscripten' build. Mostly seems to work except I get errors about:

....../microwindows/src/nanox/srvnonet.c:11:10: fatal error: 'emscripten.h' file not found

and also missing ft2build.h 

It does build to completion, but the resultant 'binary' doesnt run.

I did notice that I had to use 

emsdk install latest-fastcomp & emsdk activate latest-fastcomp

as well.

ghaerr:
Removes EMTERPRIFY options and uses only WASM code and ASYNC=1, as older EMTERPRETER has been removed from EMSCRIPTEN.

All Win32 and Nano-X programs now build and operate correctly with latest Emscripten version.

Fixes race problem with web pages not completing .js or .data file loads before program exit.

Should fix #37 and probably #55.

To build the EMSCRIPTEN port (builds all demos and runs demo-aafont):
```
cp Configs/config.emscripten config
./make-emscripten
./emrun demo-aafont
```

### DIFF
--- a/src/Configs/config.emscripten
+++ b/src/Configs/config.emscripten
@@ -227,7 +227,7 @@ HAVETEXTMODE             = N
 # SCREEN_DEPTH is bits per pixel, only used with MWPF_PALETTE palette mode
 ####################################################################
 SCREEN_PIXTYPE           = MWPF_TRUECOLORARGB
-#SCREEN_PIXTYPE           = MWPF_TRUECOLORABGR 
+#SCREEN_PIXTYPE           = MWPF_TRUECOLORABGR
 #SCREEN_PIXTYPE           = MWPF_TRUECOLOR565
 SCREEN_DEPTH             = 8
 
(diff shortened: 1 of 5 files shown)


pmp-p:
xhello.html works fine on mozilla , but it did not for me on  chrome/chromium/brave  from 87.0.4280.88 till 87.0.4280.101
could you confirm ?  i could investigate a bit if needed

ghaerr:
Hello @pmp-p,

Thanks for testing this out. I just tried it out using emscripten 2.0.11 and Chrome (unknown old version, and auto-updated to 88.0.4324.16) and xhello ran on both versions. Was it just xhello that is not working, or other demos as well?

I haven't had the time to setup the specific Chrome versions numbers your failure is occurring on, I am sorry.

Thank you!

pmp-p:
Thanks for the confirmation , so there's a problem indeed but i'm used to it and i'ill look into it :
I suspect a problem with a non-reproducible build. It happens often with emscripten + make because their clang script tend to pick up includes from build os in /usr/include  instead of staying solely in wasm sysroot and erroring that they are not found.

PR #67, opened 2022-05-13
Fix for the Atari Jaguar platform

djipi:
Fix a typo error in the Makefile_nr file.
Updated the readme file specific for the platform.


ghaerr:
Hello @djipi,

Glad to see you're still around, lurking!! :)

Thanks for the fixes. As you've seen, I'm working hard to update Makefile_nr and also keep your AJAGUAR port integrated.

If I remember, when the repo got messed up during the process of your original port, there remained outside some important changes to memory-based pixmap manipulations you wrote, something to do with using a 16-bit (or different sized) boundary. 
Do you still have those changes, you might post as a diff from back then, so that I might integrate them into the repo?

Also, is the link line for actually creating Atari executables too complicated to include (somehow) in Makefile_nr? I have been working to fix up Makefile_nr so that it is much more understandable, maintainable, but also create demo binaries.

Thank you!

djipi:
Yes, I still look at the repo time to time.

I remember the sized boundary for the pixmap, it was done for blitter operation. I think I have kept the sources, I need to check  and to make sure it was working fine of course.

About the Atari Jaguar executables, it requires specific setup (crt0, libs, etc.) and cannot be set on the Makefile_nr for my humble point of view. The console doesn't have an OS, this is bare metal. It is much simpler to have the Makefile_nr creating microwindows libs for the platform and then using specific makefile(s) to deal with the executables. Basically one rom/binary by demo.

I think a port for the Atari ST/TT/Falcon computers is doable, they have an OS. Would be fun to do it, but I do not have such systems on hand, these systems have a nice community thought.

Thank you for microwindows.

PR #95, opened 2025-02-08
Port Nano-X back to ELKS after 25 years

ghaerr:
This commit ports Nano-X back to 16-bit ELKS, which has not been supported for 25 years! ELKS can now run many of the Nano-X demos as clients talking to the ELKS-hosted server, for the first time.

Lots of changes were made, mostly to reduce the size of the client and server libraries, but also to bring back a high speed version of the VGA 4-planes driver for EGA and VGA hardware running in 640x480 16-color mode.

Changes include:
- Support ia16-elf-gcc compiler
- Reduce max request size from 30k to 1k
- Add MW_FEATURE_AREAS, MW_FEATURE_CLIENTDATA, MW_FEATURE_TINY in order to reduce size of executables
- Trim MWIF_ image formats down to 16-bits from 32-, so that switch statements work with ia16-elf-gcc
- Mouse tries /dev/ttyS0, then /dev/ttyS1 so that QEMU can be used in -serial stdio mode, with both QEMU mouse on /dev/ttyS1 and debug output to terminal on /dev/ttyS0
- Simple Makefile.elks make system instead of using "config" file (for now)

To build, the [ELKS repo](https://github.com/ghaerr/elks) must be built and the TOPDIR= environment variable set to the repo root. Then type:
```
$ make clean
$ make -f Makefile.elks clean
$ make -f Makefile.elks all
```
Binaries and libraries will be produced in bin/, lib/ and $(TOPDIR)/elkscmd/rootfs_tempate/root (/root on ELKS filesystem).
After building, create an ELKS image as usual:
```
$ cd $(TOPDIR)
$ make 
$ ./qemu.sh (must have SERIAL= uncommented for mouse support otherwise "export MOUSE_DEV=none" in /bootopts)
```
From within ELKS, the Nano-X server and simple panel launcher can be started within a shell script or using:
```
# ./nano-X & exec ./nxpanel
```


ghaerr:
Here's a screenshot of QEMU running 16-bit ELKS running Nano-X and five clients:

NX

ghaerr:
Hello @tyama501,

I have not completed all the updates necessary to get PC-98 running, but am working on it. The next step will be to add CGA support through scr_vga.c (old scr_bios.c) driver, but using CGA=1 as environment variable rather than a separate driver. Then I think we can split scr_vga.c into two parts, the shared VGA_drawhline, VGA_drawvline etc routines with PC-98 into a common file, and the VGA_open/VGA_getscreeninfo and the PC98_open/PC98_getscreeninfo in separate files. This should allow us to keep the changes between the platforms to a minimum.

I assume that the PC-98 uses the same hardware access to the 4-planar hardware as does IBM PC, it is just setup differently when going into and out of graphics modes, correct? Everything else would be the same? This also seems to be the case with the CGA 320x200 4-planar driver you wrote.

Thank you!

tyama501:
Hello @ghaerr 

Thank you.

PC-98 can access directly to 4 different plane screen address so I think it is a little different from ibm, isnit it?

https://github.com/ghaerr/elks/blob/master/elkscmd/nano-X/drivers/vgaplan4_pc98.c

For CGA driver scr_cga
I remember nxtetris works better without MONOMODE = 1 in the environment variable since it does not have mono color mode.
landmine needs MONOMODE = 1 to show lines.

toncho11:
When I try to compile it does:

~/microwindows/src# make -f Makefile.elks
ia16-elf-gcc -mcmodel=small -melks-libc -mtune=i8086 -Wall -Os -mno-segment-relocation-stuff -fno-inline -fno-builtin-printf -fno-builtin-fprintf -Wno-implicit-int -Wno-parentheses -I/root/elks/include -I/root/elks/libc/include -I/root/elks/elks/include -Wextra -Wtype-limits -Wno-unused-parameter -Wno-sign-compare -Wno-empty-body -D__ELKS__ -DELKS_VERSION=\"0.9.0-dev\" -DELKS=1 -DUNIX=1 -DNANOWM=1 -DNDEBUG=1 -DMWPIXEL_FORMAT=MWPF_PALETTE -DSCREEN_PIXTYPE=MWPF_PALETTE -Iinclude -Ielksdrivers -I/root/elks/cross/lib/gcc/ia16-elf/6.3.0/include -Wno-unused-variable -Wno-unused-but-set-variable -Wno-missing-field-initializers -DNONETWORK=0 -c -o nanox/client.o nanox/client.c
nanox/client.c:32:20: fatal error: sys/un.h: No such file or directory
 #include <sys/un.h>

So there is no unix socket support for ia16-elf-gcc?

tyama501:
Hi @toncho11 

I think you need to pull in elks repo to get libc/include/sys.h
It is added recently.

toncho11:
Thanks @tyama501 

@ghaerr 

This does not use the `TOPDIR` in the Makefile
```
bin/nano-X: $(SERVER)
        $(LD) $(LDFLAGS) -maout-heap=0x2000 -o $@ $^ $(LDLIBS)
        cp $@ /Users/greg/net/elks-gh/elkscmd/rootfs_template/root
```

toncho11:
[nano_x.zip](https://github.com/user-attachments/files/18722513/nano_x.zip)

First I think the default build is set to PC98 instead of IBM, so that is why I could not boot my images in an IBM emulator.
And I tried the above image and all I get is a mouse cursor in Box86. Mouse emulation is enabled as "Microsoft serial Emulation". Or maybe it takes too long to draw? But I am currently using pentium emulation. Actually I am getting some line in the middle of the screen from the center to the right.

toncho11:
Fuzzy line as when there is a problem with graphics.

toncho11:
Actually this fuzzy line ... is the list of applications in npanel, but because of the wrong VGA mode or something like that it becomes a fuzzy line in the center.

Also nxtetris is black and white and starts from the middle of the screen when executed:
`./nano-X & exec ./nxtetris`

So it is definitely in some CGA mode and not VGA. The Makefile.elks says:

```
CONFIG_ARCH_IBMPC=y
CONFIG_HW_VGA=y
CONFIG_ARCH_PC98=n
```
so why the wrong mode?

toncho11:
Also exiting nano-X usually exits the init and bash. So you are presented with new login.

ghaerr:
> Also nxtetris is black and white and starts from the middle of the screen when executed:

@toncho11, I agree, sounds like it is building for CGA for some reason. Lots has changed in the last few hours.

Here's a 2880k bootable image compiled for VGA with everything working.
[fd2880.img.zip](https://github.com/user-attachments/files/18722746/fd2880.img.zip)

> Also exiting nano-X usually exits the init and bash. So you are presented with new login.

I already answered that from @tyama501's same comment. Don't use 'exec' and you'll get the shell prompt back, and see the shell manual for exec.

toncho11:
All my problems are coming from the constant overriding of config.in. So I get switched to pc-98 and floppy image occasionally.

toncho11:
It all works on Pentium class socket 5 machine emulation .
On Amstrad 1640 emulation it does work, but one thing I noticed that the mouse is very slow and almost unusable when outside of an nx application. For example inside nxclock it is ok, but outside nxclock it becomes extremely delayed with like 1 or 2 seconds delay and flickering.

toncho11:
Which is the compile option for rectangle only windows while moving?

toncho11:
Yes. It is because of the mouse cursor. It is big and probably too heavy to redraw each time.  This could even explain some of the perceived slowness for other nx apps such as landmines. So there should also be an option which sets a cross cursor with just 2 lines (as it is when mouse hovering over nxclock).

toncho11:
Moving the mouse actually blocks nxclock. Once I stop moving the mouse (outside nxclock) nxclock jumps at the time it needs to be. If I am over nxclock (but not in titlebar of nxclock) nxclock and mouse work well. 
So it could also be some event handling, but only outside an nx app.

ghaerr:
> Moving the mouse actually blocks nxclock. Once I stop moving the mouse (outside nxclock) nxclock jumps at the time it needs to be. If I am over nxclock (but not in titlebar of nxclock) nxclock and mouse work well.

Thanks - I have been noticing something strange like this, but not to that detail. I have also less frequently seen cases where a mouse up after a window move doesn't actually doing anything until the next mouse event (movement).

>  It is because of the mouse cursor. It is big and probably too heavy to redraw each time. This could even explain some of the perceived slowness for other nx apps such as landmines.

Interesting, I've made a note of this also and will look into what might be done to speed things up.

> Which is the compile option for rectangle only windows while moving?

This is OUTLINE_MOVE, set in include/mwconfig.h. This is default ON for ELKS but you can turn it off to see the speed of moving windows with full contents display. It'll be slow on real hardware!

ghaerr:
> Moving the mouse actually blocks nxclock. Once I stop moving the mouse (outside nxclock) nxclock jumps at the time it needs to be. If I am over nxclock (but not in titlebar of nxclock) nxclock and mouse work well.

I can't get this to repeat - but can repeat in a somewhat opposite way: moving the cursor *inside* nxclock stops the clock from updating only while I keep moving the cursor. Should I stop moving the cursor inside nxclock, or move or not move the cursor outside, nxclock works.

Does this happen on a slow accurate emulator, a fast emulator, or real hardware? I'm only testing on QEMU.

toncho11:
So it is already in outline mode. It was so slow that it was not clear.

It is on slow emulator of Amstrad 1640, 10 Mhz. 
So the mouse is with a delay of 1 to 5 seconds. If you make a bigger movement then there is a bigger delay. So it is actually hard to even cliock on nxclock. You need to do it slowly with many small movements.

Actually you are right. The mouse might move better inside clock, but still it interferes with the nxclock updates.
So mouse movement blocks nxclock in both cases, but the speed inside nxclock becomes normal (without any delay).
If there are two nxclock they both stop while the mouse moves.

Also moving one nxclock over another nxclock freezes everything. 

Moving the mouse also stops nxtetris, but now that the cursor is the same it keeps the same delayed response to mouse movements.

ghaerr:
> So the mouse is with a delay of 1 to 5 seconds. If you make a bigger movement then there is a bigger delay. 

There's an option in the mouse driver src/drivers/mou_ser.c SLOW_CPU that can be set to drop mouse events preventing a bigger delay. That could help. I haven't yet had time to try to see what things can be done for operating on a super slow machine, but likely not too much. Another option which would likely be better would be to discard all previous mouse move events when a new one is read, I'll look into adding that with some testing.

> Also moving one nxclock over another nxclock freezes everything.

That should not crash the system, but it does cause the number of "clip rectangles" to be generated and searched to grow and thus slows down drawing even further. I'm hoping that is what is happening. (There is a reason why Windows 1.0 had tiled-only windows!!)

> maybe mouse pointer is too heavy and has a very delayed respons

Probably true, will try eating old mouse events to see if that helps.

> mouse movements block all nx apps

I can repeat that if the mouse is moved inside nxclock, nothing happens for the duration of the mouse movement. I'll try fixing that somehow.

Thanks for your testing! Please report bugs on full-speed emulator also if you have a chance, as I am seeing some glitches on occasion even on fast systems.

ghaerr:
Progress is being made, and a port of the very cool updated nxterm is now working well. There are still a few drawing issues,  the complexity of which are causing long chases down rabbit holes. Almost all of today was spent on them. I was hoping to have more done but there's still just too many issues to commit anything yet.

Here's a screenshot of nxterm running /bin/sh and vi - it is working quite well except for a few out-of-window blit/clip errors (look carefully just to the left of the nxterm window in the first screenshot). Nxtern features a full ANSI emulation with forward/reverse scrolling, blinking cursor, etc, much improved from the present ELKS version. The full ASH shell is also running and the nxterm functions just like the ELKS console, with most of its capabilities.

I'm striving for a bit of perfection with the graphics draw code, and soon we will be at a point where ELKS could be said to have full graphics capabilities and a real desktop of sorts - or at least a great terminal emulator and some graphics games. (BTW, there is a Doom ported to Nano-X, but its a 32-bit version). 

Overall, the operation of all each of the programs running in client/server on NX seems to be quite stable, at least until the system resources are used up, which can easily happen after 5-6 windows open. I'm pretty impressed with the system!
nxterm ls

nxterm

@toncho11, I've also duplicated the mouse problems you've brought up, they're also proving to be complicated, still working on them.

toncho11:
Impressive! :)

ghaerr:
> I've also duplicated the mouse problems you've brought up, they're also proving to be complicated, still working on them.

With PR #102, a smaller cursor was created which greatly reduces the CPU time spent just moving the cursor on the screen. Also, moving the cursor doesn't stop other applications from running, or at least it shouldn't except on possibly very slow systems. Overall, the system speed was sped up greatly on slow systems. 

Nxterm however is *very* slow when scrolling, and about the only solution to that is to decrease the terminal width and height from 80x25 to something lots smaller. This affects ls -l output so its not a great solution either.

PR #96, opened 2025-02-08
First pass at IBM CGA and PC-98 screen drivers

ghaerr:
All screen, mouse and keyboard drivers are converted for IBM/CGA and PC98 platforms and are compiling without warnings.

I haven't had time to test PC-98, but IBM CGA ran in black and white but the internal screen height/width was incorrect, as 640x480 is hard-coded for ELKS in include/mwconfig.h for the time being. I haven't thought of a way to configure that easily, we aren't using the standard "config" method but a simpler one to build - see below.

@tyama501, I tried to keep all of your original driver code for CGA and PC98 without modifications, but the new Nano-X driver format is quite a bit different. I think we are close! :)

In order to build, the following lines in Makefile.elks needs to be edited as desired:
```
# Modify the following lines to build IBM/VGA, IBM/CGA or PC98 drivers
CONFIG_ARCH_IBMPC=y
CONFIG_HW_VGA=n
CONFIG_ARCH_PC98=n
```
and then run 
```
make clean
make -f Makefile.elks clean
make -f Makefile.elks all
```

### DIFF
--- a/src/Makefile.elks
+++ b/src/Makefile.elks
@@ -10,6 +10,11 @@ include $(BASEDIR)/Makefile-rules
 # Uncomment the following line to link nano-X applications with the server.
 #LINK_APP_INTO_SERVER=1
 
+# Modify the following lines to build IBM/VGA, IBM/CGA or PC98 drivers
+CONFIG_ARCH_IBMPC=y
+CONFIG_HW_VGA=n
+CONFIG_ARCH_PC98=n
+
 # remove any -mregparmcall as ASM files need cdecl calling convention
 CFLAGS := $(filter-out -mregparmcall, $(CFLAGS))
 CFLAGS += -DELKS=1 -DUNIX=1 -DNANOWM=1 -DNDEBUG=1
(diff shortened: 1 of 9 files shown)


tyama501:
Hello @ghaerr 

I have tried on PC-98 emu.

It seems that draw position has offset.
Also the mouse is not working.

Esc can exit, but it goes to login:  (is this also occur in you environment?)

./nano-X & exec ./npanel
 
image

./nano-X & exec ./nxworld
(a lot slower than built in server...)

image

I will take a look when I have more time.
Thank you.

BTW, in the makefile somehow I needed to comment out below.
-ifndef $(TOPDIR)
-TOPDIR = /Users/greg/net/elks-gh
-endif
+#ifndef $(TOPDIR)
+#TOPDIR = /Users/greg/net/elks-gh
+#endif

-       cp $@ /Users/greg/net/elks-gh/elkscmd/rootfs_template/root
+       cp $@ $(TOPDIR)/elkscmd/rootfs_template/root
+#      cp $@ /Users/greg/net/elks-gh/elkscmd/rootfs_template/roo

tyama501:
Well, not bad for the first try!
Thank you for your great effort!

toncho11:
> +#ifndef $(TOPDIR)
> +#TOPDIR = /Users/greg/net/elks-gh
> +#endif

I also had to comment this or set it to my own folder.
This check does not seem to work and it sets the folder /Users/greg/net/elks-gh

ghaerr:
Thanks @toncho11, I've just fixed both these issues in #97.

PR #97, opened 2025-02-08
Various VGA, CGA and PC98 driver fixups

ghaerr:
Fixes off-by-one vertical line bugs seen in https://github.com/ghaerr/microwindows/pull/96.

More cleanups coming.

### DIFF
--- a/src/Makefile.elks
+++ b/src/Makefile.elks
@@ -1,9 +1,6 @@
 # Nano-X Makefile for ELKS
 # originally from v0.86 10/29/99 g haerr
 
-ifndef $(TOPDIR)
-TOPDIR = /Users/greg/net/elks-gh
-endif
 BASEDIR = $(TOPDIR)/elkscmd
 include $(BASEDIR)/Makefile-rules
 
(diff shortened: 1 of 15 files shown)


toncho11:
Please commit this PR so that we can test it easier.

ghaerr:
Ok, I'll commit now and continue with another PR.

The VGA driver should be working well, CGA is better but drawing off the screen. PC-98 should be in better shape but haven't yet added the GsCheckMouseEvent required for mouse operation.

@tyama501, are you testing this on a HD image for PC-98 on DosBoxX (or Neko)? The 1440k image is completely out of space, and I'm almost out on 2880k. I'm not sure how to setup a larger image, or whether it just works if you pass it one. Also wondering if there's anything special about getting the mouse setup, or if that is automatic on both emulators.

tyama501:
Thank you!

I haven't checked this PR yet,
but yesterday I have used fd1232 image since
I have been leaving space that can fit new nano-X.

I had used Neko but it should work with Dosbox-X too.
Neko set up mouse automatically.
Dosbox-X may need to drop down list and do Capture mouse.

PR #101, opened 2025-02-09
Speed up screen drivers, decrease caption height

ghaerr:
A hardware color is passed to the screen driver as MWPIXELVAL, which is 32-bits, so long arithmetic was being used in all the VGA screen drivers for manipulating color, resulting in lots of slowdown. They're fast now, but a better solution would be using MWPIXELVALHW for driver color parameters.

Reduces title bar by 4 pixels for ELKS which displays better on smaller 640x480 screens.

`npanel` was renamed `nxpanel` so startup now is:
```
./nano-X & ./nxpanel
```

Here's a bootable 2880k image and screenshot:
[fd2880.img.zip](https://github.com/user-attachments/files/18722935/fd2880.img.zip)

NX smaller caption

### DIFF
--- a/src/Makefile.elks
+++ b/src/Makefile.elks
@@ -59,8 +59,7 @@ NANOX = nanox/srvmain.o nanox/srvfunc.o nanox/srvutil.o nanox/srvevent.o \
 NANOX += nanox/wmaction.o nanox/wmclients.o nanox/wmevents.o nanox/wmutil.o
 ENGINE = engine/devdraw.o engine/devmouse.o engine/devkbd.o engine/devclip1.o \
 	engine/devopen.o engine/devfont.o engine/devlist.o engine/devblit.o \
-    engine/devtimer.o \
-    engine/devpal1.o engine/devpal2.o engine/devpal4.o engine/devpal8.o
+    engine/devtimer.o engine/devpal1.o engine/devpal4.o
 
 ALL = lib/libnano-X.a nxdemos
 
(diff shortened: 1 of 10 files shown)


tyama501:
Hello @ghaerr 

nxpanel on the pc-98 looks better than before. There is no line offset anymore. Thank you.
image

It seems that  the mouse is still not working.
Maybe I will try to debug with printf later since PC-98 can use text console :)

tyama501:
It seems that GsSelect in the srvmain is never called after GrOpen.
image

ghaerr:
> It seems that GsSelect in the srvmain is never called after GrOpen.

Oh, of course! We're not compiling under ELKS, so "CONFIG_ARCH_PC98" is not defined! For now, you will have to add it manually to Makefile.elks using CFLAGS += -DCONFIG_ARCH_PC98=1. Meanwhile I will fix this in the Nano-X repo in my next commit.

> nxpanel on the pc-98 looks better than before.

Yes, it looks good. I do notice that the line directly under the window title bar is yellow, instead of red on my system. Do you know if the standard PC-98 palette is the same on PC98 as on IBM PC? If not, we will need to setup a separate palette table.
(I don't know if I like either yellow or red color in either of these cases, but for now have left it since many other problems to deal with. All colors are now specified in RGB but using a "min distance" algorithm to compute a 4-bit color from 24-bit RGB).

tyama501:
It is different but I use correct plane to get similar colors for 0 to 7.
Maybe colors for 8 to 15 are different since there is no brown.

Most PC-98 can choose 16 colors out of 4096 colors (12bits) by IO control.

Thank you.

PR #102, opened 2025-02-12
Port nxterm to ELKS and lots of bug fixes

ghaerr:
Nxterm is now available and running on ELKS! 

Creating this PR took quite some time because quite a few other issues were found, particularly in the VGA drivers, that took forever to solve. There remain a number of issues and desired enhancements to nxterm, but progress overall looks good for the Nano-X ELKS port, and this PR also fixes a number of problems seen operating on slow systems.

Most testing was done on QEMU by running the Nano-X panel program launcher after building and booting ELKS using:

Building Nano-X for ELKS
```
$ make clean
$ make -f Makefile.elks clean
$ make -f Makefile.elks
$ cd $TOPDIR
$ make
```
Running Nano-X within ELKS
```
login: root
# ./nano-X & ./npanel
```

Features added and bugs fixed for ELKS:
- Nxterm port. Lots of fixes for all platforms, including non-buggy up/down scrolling - enough to run ELKS vi. Still needs some ANSI reverse scrolling fixes.
- Exit Nano-X on ^A instead of ESC. This was required because vi uses ESC.
- Desktop color changed from CYAN to GRAY for more modern look on 16-color VGA.
- Add USE_SMALL_CURSOR for small 8x8, rather than 16x16 cursor. Lots of CPU time is spent just redrawing the cursor and having a cursor 1/4 the size makes quite a bit a difference.
- Nano-X event fixes: some window move starts were ignored, and sometimes mouse events weren't processed until next event occurred (e.g. end of window move not acknowledged properly and cursor not changing after window move).
- VGA blit routine bug after XOR used. Blit still improperly draws outside the application on scrolling. 
- VGA speedups on horizontal line draw and draw pixel.
- More size reduction of Nano-X client/server libraries for ELKS.
- Fixes issues brought up in https://github.com/ghaerr/microwindows/pull/95#issuecomment-2646575057 by @toncho11.

Known bugs:
- Window contents sometimes not refreshed when clicking on titlebar.
- Screen contents just outside the left or right side of nxterm is overwritten when scrolling due to VGA planes blit bug.

Needs enhancements:
- Nxterm is very slow scrolling on slow systems!!
- Nxterm doesn't redisplay contents after being obscured.
- Nano-X redisplays application background twice on startup (because of window manager caption drawn).
- Nxclock time display can look funny when app display becomes unobscured (because of XOR drawing mode).

For those without time to build ELKS, attached is a bootable 2880k floppy image with Nano-X ready to run as described above:
[fd2880.img.zip](https://github.com/user-attachments/files/18762202/fd2880.img.zip)

Here's a screenshot showing the (optional) small cursor used for ELKS (to the right of nxterm window):
NX cursor

### DIFF
--- a/src/Makefile.elks
+++ b/src/Makefile.elks
@@ -14,8 +14,9 @@ CONFIG_ARCH_PC98=n
 
 # remove any -mregparmcall as ASM files need cdecl calling convention
 CFLAGS := $(filter-out -mregparmcall, $(CFLAGS))
-CFLAGS += -DELKS=1 -DUNIX=1 -DNANOWM=1 -DNDEBUG=1
+CFLAGS += -DELKS=1 -DUNIX=1 -DNDEBUG=1
 CFLAGS += -DMWPIXEL_FORMAT=MWPF_PALETTE -DSCREEN_PIXTYPE=MWPF_PALETTE
+CFLAGS += -DUSE_SMALL_CURSOR=1
 CFLAGS += -Iinclude -Ielksdrivers -I$(TOPDIR)/cross/lib/gcc/ia16-elf/6.3.0/include
 CFLAGS += -Wno-unused-variable -Wno-unused-but-set-variable
 CFLAGS += -Wno-missing-field-initializers
@@ -23,6 +24,7 @@ AR = ia16-elf-ar
 
 # screen, mouse and kbd drivers
 ifeq ($(CONFIG_ARCH_PC98), y)
+CFLAGS += -DCONFIG_ARCH_PC98=1
 DRIVERS += drivers/scr_pc98.o drivers/vgaplan4_pc98.o drivers/vgaplan4_mem.o
 DRIVERS += drivers/mou_pc98.o drivers/kbd_tty.o
 #DRIVERS += drivers/ramfont.o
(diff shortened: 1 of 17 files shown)


rofl0r:
you might wanna have a look at uuterm that has super fast rendering. maybe there's a trick or two in it you could copy. it should be even be quite easy to have it use nx instead of libx11 http://git.musl-libc.org/cgit/uuterm/tree/README

ghaerr:
Thanks @rofl0r. I've looked at uuterm before and it is pretty cool - thank you for the reminder I'll take another look now that I'm way more than knee deep into nxterm. I'm thinking of using a "jumpscroll" technique where unprocessed input linefeeds are counted and a single scroll up for that amount is done once rather than each LF seen.

toncho11:
Excellent job!
Now I can use the mouse on an Amstrad!
nxtetris starts slowly. It seems it does multiple repaints until a new game is possible.
Good job! :)

ghaerr:
> Now I can use the mouse on an Amstrad!

Thanks for testing, I'm glad to know NX now runs on your Amstrad. I was especially surprised chasing down the speed issue, only to find that there weren't any bugs and that a 16x16 cursor was just too large to process. 

> nxtetris starts slowly. 

Are you talking about the two repaints, or is running slowly in other ways? The double repaint problem is an artifact of the original NX then a window manager was added. What is happening is that NX draws the original window, then the WM gets notified and then sends a window move request in order that the title bar can be placed in the original X,Y position. This causes a flicker at best and on slow systems one can see the entire paint/erase/repaint sequence. I'm working on a solution but don't have one yet.

toncho11:
nxpanel dekstop shows much more evidently that ELSK can multi-task.

PR #106, opened 2025-02-14
Add nxstart to show Nano-X demos

ghaerr:
Lots of small enhancements made to more easily allow a user to play around with Nano-X demos.

The "npanel" program was renamed "nxstart", and can now be used to run a lot of the other demos. 
Also, client programs now automatically start the Nano-X server, so running shell script or "nano-X &" etc are no longer required.

In order to start the Nano-X demo, just type "bin/nxstart" (on ELKS images "./nxstart").

Other changes include:
- Add lots of demos to nxstart
- Rename many Nano-X demo binaries to start with "nx"
- Fix nxtetris slow startup reported by @toncho11
- Get nxlaunch running as a desktop alternative to nxstart (not available on ELKS yet)
- Standardize using GrNewWindowEx(GR_WM_PROPS_APPWINDOW, ...) for most demos

Here's a demo screen on macOS:

nxstart

Prebuilt ELKS 2880k floppy image:

[fd2880.img.zip](https://github.com/user-attachments/files/18791746/fd2880.img.zip)


toncho11:
Great. It will be interesting to se the calculator and Chess on ELKS.
I have a problem mouse though. I think before Microsoft Serial Mouse on COM1 was working. No there is no mouse support.
Actually `mouse` program reports movement, so it must be nano-X.

ghaerr:
> I have a problem with mouse though. I think before Microsoft Serial Mouse on COM1 was working. No there is no mouse support.

Are you running on an emulator or real hardware? The mouse support has got a bit more complicated with QEMU support.
You might try uncommenting the MOUSE_PORT=/dev/ttyS1 (or ttyS0) line in /bootopts.

> Actually mouse program reports movement, so it must be nano-X.

That is strange, although the drivers are different, they're pretty much equivalent. I

> It will be interesting to se the calculator and chess on ELKS.

The calculator is built to demonstrate Nuklear, a "small" immediate-mode UI whose .h file is 98k, so It likely won't be running on ELKS anytime soon. I'll have to check chess, it's built out of several source files and was omitted from the ELKS build because of the current directory structure.

tyama501:
Hello @ghaerr ,

It worked very well on PC-98!!
(Although there is no enough heap for all 5 applications)

This is 12MHz setting. Amazing. Thank you.

https://github.com/user-attachments/assets/5f93e4b7-23b7-48b4-8d49-d75862adee2b

ghaerr:
> It worked very well on PC-98!!
> This is 12MHz setting. Amazing. Thank you.

Wow, that is amazing considering that it wasn't tested!! Thank you for your testing. Now that this is working, please play around with it more and see what you think. In particular, look for any drawing issues or incorrect colors that might need to be fixed. I am already aware of the yellow bar just underneath the Start menu.

This new version supports lots more in terms of fonts, although most of that is currently turned off due to keeping initial size down. But getting Japanese output and/or Japanese NX applications will be much easier with the new support already built in to Nano-X.

> Although there is no enough heap for all 5 applications

This should be easily fixed by changing the linker heap to nano-X on line 84 of Makefile.elks and recompiling. Only nano-X itself should need expanded heap. I kept the value at 8K (0x2000) to leave the most room for other applications, because memory is still extremely tight.

I am very glad to hear that we have successfully transitioned to the new Nano-X on both IBM and PC-98 platforms!

BTW, does it seem faster? I fixed a number of speed issues and turned asserts off with -DNDEBUG=1. There are probably areas in the PC-98 screen driver that could also be sped up but I tried to fix most of them already.

ghaerr:
@tyama501: if the timing is off on nxtetris, you may want to delete the following lines in nanox/srvmain.c to see whether they are still needed or not. I would think they are no longer needed, and they also screw up the timing for programs like nxtetris:
```
#if CONFIG_ARCH_PC98
    if (timeout == GR_TIMEOUT_BLOCK)
        timeout = 10;
#endif
```
Let me know if these lines can be deleted and I will fix it over here.

tyama501:
Thank you @ghaerr 

I think the video dropped some frame while recording. It is a little more smooth.

I think the brown is changing to yellow since
color code 6 is brown for cga but it is yellow for PC-98.

ghaerr:
> I think the brown is changing to yellow since
color code 6 is brown for cga but it is yellow for PC-98.

The color scheme is now specified in RGB, then converted to a standard (IBM) 16-color palette through GdFindNearestColor(). You can play with the color scheme by looking at the file nanox/nxdraw.c, under the scheme SCHEME_NUK16:
```
    /* window border and interior line under caption*/
    GR_RGB( 32,  32,  32),  /* GR_COLOR_WINDOWFRAME         */
    GR_RGB( 65,  65,  65)   /* GR_COLOR_WINDOWFRAMELT       */
```
In particular, I think the last line GR_COLOR_WINDOWFRAMELT could possibly be changed to produce a different value for PC98, if wanted. This may or may not be better than changing the default palette or changing GdFindNearestColor. You can look further into this after testing that Nano-X is actually working well on PC-98.

toncho11:
Yes, it is faster I think. 
The mouse problem persists. It seems nano-X sees the mouse because it does load, but when I try to move it - nothing happens. I am using it on 86Box emulator. The `mouse `program should say where it detected a mouse.

toncho11:
I tested for a while but I could not figure it out. Sometimes ./nxstart won't start without any complain. I changed on which port is the mouse on the emulator, but I could never make it work.

On which com port should be the mouse by default if nothing is set in /bootopts?
tty0 is the console and not com1?
There is some confusion here.

ghaerr:
> I am using it on 86Box emulator.

A change was made which should not affect you, but it seems it is. The issue is which device should the mouse should be on, and what was added was the capability to display "error" or "printf" output from an emulator, and which device that is mapped to *inside* the emulator.

It used to be the mouse was on /dev/ttyS0, which is still the case for real hardware. But then on QEMU there is no way to "map" debug output to the host window (not inside ELKS, but the host window the emulator is being run from). Long story short, this has to be /dev/ttyS0 for QEMU, which then means the mouse has to move to /dev/ttyS1.

What was changed is that the mouse open code looks to see if running on QEMU by checking the QEMU=1 environment variable. If this is set, then the mouse driver opens /dev/ttyS1 for the mouse. On QEMU, one can see the debug messages since they are written to /dev/ttyS0, which can be the ELKS console.

On real hardware, the mouse driver tries /dev/ttyS0 first, then, if that can't be opened, /dev/ttyS1. The debug output will still try to go to /dev/ttyS0, otherwise /dev/console. If all this sounds confusing, it kind of is.

I don't know how you have configured 86Box and I don't have it installed. Perhaps try setting `export QEMU=1` in the shell before running ./nxstart to see what happens. Or try setting up the mouse to be at com2 in 86Box settings and set MOUSE_PORT=/dev/ttyS1 in /bootopts.

Once we figure out what the 86Box issue is, I can sort more things out, but I can't change back the mouse to com1 default without losing the ability to see debug output which is crucial. That said, I can make a change to output which mouse port is being used, but that is typically erased from the ELKS console as soon as it is displayed and goes into graphics mode.

> The mouse program should say where it detected a mouse.

I made that change with the last commit. But "mouse" operates a bit differently than nano-X, and doesn't use QEMU=. It always tries to use /dev/ttyS0 unless MOUSE_PORT=/dev/ttyS1.

Summary: try setting up the mouse to be at com2 in 86Box settings and set MOUSE_PORT=/dev/ttyS1 in /bootopts.
Also try to setup 86Box to display any output to /dev/ttyS0 back to the host terminal if possible.

toncho11:
Long story short ... this sounds complicated :)
I do not see QEMU variable after boot and using printenv.
There is no QEMU on /bootopts neither (there was before).
I tried different configurations as you suggested, but I could not make it work.

ghaerr:
I agree my explanation is complicated but I can't seem to figure out what is failing. It sounds like either the mouse is on /dev/ttyS1 but NX thinks its on ttyS0 or vice versa. The only way to force it is to use MOUSE_PORT= in /bootopts or `export MOUSE_PORT=/dev/ttyX` in the shell.

Unfortunately, there is no way to display a debug message since the console is in graphics mode!

ghaerr:
You say `mouse` is working? Is that with or without a MOUSE_PORT= in effect?

I will add a change to `mouse` to display the mouse port being used, that will help let us know what is happening.

tyama501:
Hello @ghaerr ,

These are just about PC-98 but I have noticed two things.

The cursor on the text console is blinking at the left bottom corner.
I know it can be disabled by ESC[?25l prior to nxstart, but is there way to disable this when initializing the graphic driver
and enabling when exit?

One more thing,
is the CTRL-C only way to exit to console?
It seems that PC-98 graphic mode is not disabled in this exit,
so, the last image still shows after exit to the console.

Thank you.

toncho11:
CTRL + A to exit nano-X

tyama501:
I see. Thank you!

ghaerr:
Hello @tyama501,

> The cursor on the text console is blinking at the left bottom corner.
I know it can be disabled by ESC[?25l prior to nxstart, but is there way to disable this when initializing the graphic driver
and enabling when exit?

Yes, the text cursor will have to be disabled at startup in the drivers/scr_pc98.c::PC98_open() routine, and then enabled again in PC98_close(). You should be able to do this using the same INT 0x18h code that is being used in the ELKS console code:
```
cursor_on:
        mov $17,%ah
        int $0x18
        ret

cursor_off:
        mov $18,%ah
        int $0x18
        ret
```
However, in this case since scr_pc98.c is already using inline ASM, just add this functionality directly into the .c file using \_\_asm\_\_, rather than adding an additional .S file.

> It seems that PC-98 graphic mode is not disabled in this exit,
so, the last image still shows after exit to the console.

Any other text screen clearing can be added to the PC98 screen driver open or close routine to fix this.

After you have that working to your liking, please submit a PR :)

Thank you!

PR #108, opened 2025-02-16
Update ELKS paths in nxstart, partial fix for closebox

ghaerr:
Nano-X binaries are now copied to ELKS /bin, rather than /root. 

Since nxstart is not in /bin, to start Nano-X use:
```
# nxstart
```
without having to run ./nxstart as before.

Partially fixes the appearance of a shell hang on exit from nxstart, as reported by @toncho11 in https://github.com/ghaerr/elks/pull/2228#issuecomment-2661330717. This turns out to be an issue with the shell having problems reading the terminal that was reset to RAW mode by nano-X. Typing ^C after exit will bring the prompt back.

This PR partially fixes the appeared hang on exit from nxstart - which will now work but only when nxstart is the last window to be closed in Nano-X. A long term fix will come later.

### DIFF
--- a/src/Makefile.elks
+++ b/src/Makefile.elks
@@ -82,7 +82,7 @@ lib/libnano-X.a: $(CLIENT)
 
 bin/nano-X: $(SERVER)
 	$(LD) $(LDFLAGS) -maout-heap=0x2000 -o $@ $^ $(LDLIBS)
-	cp $@ $(TOPDIR)/elkscmd/rootfs_template/root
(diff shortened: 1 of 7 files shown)


toncho11:
I see. This is not /bin but /root/bin ?

ghaerr:
> This is not /bin but /root/bin ?

No - everything was moved out of /root to clean it up, and only binaries copied into /bin for now. See https://github.com/ghaerr/elks/issues/2157#issuecomment-2661580052.

PR #112, opened 2025-02-22
Fixed warnings on Nxchess example

prajoshpremdas:
Nxchess example reported a warning on ftime function as its depreciated and suggested usage of gettimeofday function.

Signed-off-by: Prajosh Premdas <premdas.prajosh@gmail.com>

### DIFF
--- a/src/demos/tuxchess/main.c
+++ b/src/demos/tuxchess/main.c
@@ -10,7 +10,7 @@
 #if RTEMS
 #include <time.h>
 #else
-#include <sys/timeb.h>
+#include <sys/time.h>
 #endif
 #include "defs.h"
 #include "data.h"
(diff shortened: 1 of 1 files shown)


ghaerr:
Thank you for the enhancement!

toncho11:
What are the chances on building nxchess on ELKS?

toncho11:
Maybe the image routines developed by @rafael2k is exactly what is needed to display the images.

ghaerr:
> Maybe the image routines developed by @rafael2k is exactly what is needed to display the images.

Well, it would possibly be a start, since the elks-viewer routines decode directly to the VGA display, if I'm not mistaken. However, there would still be potentially a lot of work to be done, as the entire mechanism may need reprogramming to fit within the architecture of the image_* engine routine API.

However, the biggest problem is that the Microwindows engine image loading routines assume that the entire image will be stored in Nano-X's memory, which is no problem for 32-bit machines, but likely too small except for the smallest of images for ELKS. (Remember that Nano-X itself is limited to a mere 64k total memory for *all* connected applications!) This is because Nano-X doesn't always directly decode to a screen, but always decodes to an internal buffer. Thus, a new design of some kind is probably needed in order to display images on ELKS Nano-X. I will have to dive deeper back into this to figure out what kind of design that might be.

Issue #116, opened 2025-03-03
Nuklear image png not work


big-vl:
This is probably what is now blocking in the implementation of the project, but is it possible to find out how to display Nano-X in existing ones with Nuklear, I understand that the full implementation will take a lot of time through stb_image, but I would have been held right on top of the Nuklear with the possibility of positioning.

```
#include "nx_img.h"
#include "nuklear.h"  /* For nk_image, nk_image_ptr(), nk_image_id() */
#include <stdlib.h>
#include <string.h>
#include <nano-X.h>

#define STB_IMAGE_IMPLEMENTATION
#include "stb_image.h"

/* Debug macro always enabled */
#define NX_DBG(fmt, ...) fprintf(stderr, fmt, ##__VA_ARGS__)

/*
 * Internal structure stored in nk_image.handle.ptr.
 * Contains the Nano-X pixmap (GR_WINDOW_ID) and image dimensions.
 */
typedef struct {
    GR_WINDOW_ID pixmap;
    int w, h;
} NxImgData;

/*
 * Creates an nk_image from an array of RGBA pixels.
 * On success, returns an nk_image whose handle.ptr points to a NxImgData structure.
 */
static struct nk_image nx_img_create_nk_image(const unsigned char *pixels, int w, int h)
{
    struct nk_image result;
    NxImgData *d = (NxImgData*)calloc(1, sizeof(NxImgData));
    if (!d) {
        NX_DBG("nx_img_create_nk_image: calloc failed\n");
        return nk_image_id(0);
    }

    d->pixmap = GrNewPixmapFromData(w, h,
                                    0xFFFFFFFF, /* foreground = white */
                                    0xFF000000, /* background = black */
                                    (void*)pixels,
                                    MWPF_TRUECOLORARGB);
    if (!d->pixmap) {
        NX_DBG("nx_img_create_nk_image: GrNewPixmapFromData failed\n");
        free(d);
        return nk_image_id(0);
    }
    d->w = w;
    d->h = h;

    result = nk_image_ptr(d);
    result.w = w;
    result.h = h;
    NX_DBG("nx_img_create_nk_image: created image %dx%d, pixmap id = %d\n", w, h, d->pixmap);
    return result;
}

/*
 * Loads a PNG image using stb_image and creates an nk_image.
 */
NK_API struct nk_image nx_img_load_png(const char *filename)
{
    int x, y, n;
    const int req_channels = 4; /* Request RGBA */
    
[...]

PR #117, opened 2025-03-04
Add image loading for Nuklear UI applications

ghaerr:
Adds enhancements by @big-vl from #116 to enable image loading and display in Nuklear. 

This is enabled by defining NK_NANOX_INCLUDE_IMAGE and uses the Nano-X server GrLoadImage routines (preferred).
The older STB image loader (requires X11) is enabled by defining NK_NANOX_INCLUDE_STB_IMAGE.

Requested fonts default to 14x14 and may require FreeType 2.10.4 or later, unless builtin system fonts used.
Text strings default to UTF-8 format.

Optional UTF-8 to UCS-2 conversion for Nuklear added in demos/nuklear/alt_nk_draw_text.c.

### DIFF
--- a/src/demos/nuklear/nuklear_nxlib.h
+++ b/src/demos/nuklear/nuklear_nxlib.h
@@ -602,18 +602,12 @@ NK_API void nk_nxsurf_draw_image(NXSurface *surf,
                                  struct nk_image img,
                                  struct nk_color col)
 {
-    (void)col;  /* 'col' не используется в этой реализации */
-
-    printf("nk_nxsurf_draw_image: drawing image on window id %d at (%d, %d) size (%d x %d)\n",
-           surf->wid, x, y, w, h);
+    (void)col;
 
     GR_IMAGE_ID image_id = (GR_IMAGE_ID)img.handle.id;
-    if (!image_id) {
-        printf("nk_nxsurf_draw_image: image_id is 0\n");
+    if (!image_id)
         return;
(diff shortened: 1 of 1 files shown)


src/demos/nuklear/nuklear_nxlib.h
```diff
@@ -685,7 +760,7 @@ nk_nxlib_create_window(struct nk_context *ctx)
 	title = ctx->current->name_string;
 
 	/* window w/h decreased by 1 because Nuklear draws on surface one less than size given*/
-	wid = GrNewBufferedWindow(GR_WM_PROPS_APPWINDOW, title, GR_ROOT_WINDOW_ID,
+	wid = GrNewBufferedWindow(GR_WM_PROPS_NOAUTOMOVE, title, GR_ROOT_WINDOW_ID,
```
big-vl, line 763: GR_WM_PROPS_APPWINDOW
It is better to leave it because some users may confuse this and the window will not move.

ghaerr: Yes, I saw that. Note that I changed this to GR_WM_PROPS_NOAUTOMOVE, which is a bit different than GR_WM_PROPS_NOMOVE. The former stops the window manager from changing the window position specified in the GrNewWindowEx call, while the latter stops the window from being drag-moved by the window manager. 

So everything should work as I have changed it. Thank you.

PR #121, opened 2025-03-12
Add MW_FEATURE_BITMAPS to enable GrBitmap support

ghaerr:
Required by @tyama501 for Japanese text viewer discussed in https://github.com/ghaerr/elks/discussions/1387#discussioncomment-12291997.

### DIFF
--- a/src/engine/devdraw.c
+++ b/src/engine/devdraw.c
@@ -848,7 +848,7 @@ GdDrawImageByPoint(PSD psd, MWCOORD x, MWCOORD y, PMWIMAGEHDR pimage)
 	}
 	extra = pimage->pitch - linesize;
 
-#if MW_FEATURE_IMAGES
+/*#if MW_FEATURE_ALPHA */
 	/* Image format in RGB rather than BGR byte order?*/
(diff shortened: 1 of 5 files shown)


src/engine/devdraw.c
```diff
@@ -1236,7 +1237,9 @@ GdDrawImageByPoint(PSD psd, MWCOORD x, MWCOORD y, PMWIMAGEHDR pimage)
 
 	GdFixCursor(psd);
 }
+#endif /* MW_FEATURE_IMAGES*/
```
rofl0r, line 1240: the corresponding `#if` was commented out, so either this endif is completely dangling or at least the comment no longer accurate.

ghaerr: Thanks for looking at this. I think the diff combined with my change makes this looks odd - as there used to be an embedded \#if MW_FEATURE_IMAGES on line 851 which I changed to \#if MW_FEATURE_ALPHA and commented out.

So, this \#endif and comment are actually correct and match the corresponding \#if MW_FEATURE_IMAGES on line 533.

PR #123, opened 2025-04-10
Speed up mouse cursor and VGA display for slow 8088 systems

ghaerr:
Brings over some of the changes made for ELKS Paint in https://github.com/ghaerr/elks/pull/2277 to speed up drawing.

The mouse cursor show/hide routines now only save and redraw bits set in the cursor mask. This is much faster since each 16x16 or 8x8 cursor pixel is drawn individually using drawpixel.

C constructs like x/8 or y*80 are replaced with left/right shifts in the VGA screen driver. This is because the compiler used, ia16-elf-gcc, only works well when optimizing for size (-Os) and as such emits DIV and MUL instructions for the examples respectively, which are very slow on 8088. When -Os is set, unfortunately even certain left shifts (such as (y<<6) + (y<<4)) are actually replaced ("optimized") with y * 80 and a MUL still produced. Nonetheless, the C code is changed now until a workaround for optimization level 3 (-O3) is found that doesn't produce faulty code.

### DIFF
--- a/src/Makefile.elks
+++ b/src/Makefile.elks
@@ -72,7 +72,7 @@ NANOX = nanox/srvmain.o nanox/srvfunc.o nanox/srvutil.o nanox/srvevent.o \
 NANOX += nanox/wmaction.o nanox/wmclients.o nanox/wmevents.o nanox/wmutil.o
 ENGINE = engine/devdraw.o engine/devmouse.o engine/devkbd.o engine/devclip1.o \
 	engine/devopen.o engine/devfont.o engine/devlist.o engine/devblit.o \
-    engine/devtimer.o engine/devpal1.o engine/devpal4.o
+    engine/devpal1.o engine/devpal4.o
 
 ALL = lib/libnano-X.a nxdemos
 
--- a/src/Makefile.elks
+++ b/src/Makefile.elks
@@ -14,6 +14,17 @@ CONFIG_ARCH_PC98=n
 
 # remove any -mregparmcall as ASM files need cdecl calling convention
 CFLAGS := $(filter-out -mregparmcall, $(CFLAGS))
+CFLAGS += -O2
+CFLAGS += -fno-align-jumps
+CFLAGS += -fno-align-functions
+CFLAGS += -fno-align-loops
+CFLAGS += -fno-align-labels
+CFLAGS += -fira-region=one
+CFLAGS += -fira-hoist-pressure
+CFLAGS += -freorder-blocks-algorithm=simple
+CFLAGS += -fno-reorder-blocks
+CFLAGS += -fno-prefetch-loop-arrays
+CFLAGS += -fno-tree-ch
 CFLAGS += -DELKS=1 -DUNIX=1 -DNDEBUG=1
 CFLAGS += -DMWPIXEL_FORMAT=MWPF_PALETTE -DSCREEN_PIXTYPE=MWPF_PALETTE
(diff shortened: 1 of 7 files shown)


ghaerr:
Very strangely, removing MW_FEATURE_TIMERS fixes the very slow cursor seen after running nxclock using nxstart under QEMU. This is most strange because the cursor remains slow even after Nano-X is exited an started again. When MW_FEATURES_TIMERS is unset, none of this is seen. To be investigated.

rofl0r:
at musl we've come up with a combination of CFLAGS that uses the -O2 baseline with selected optimizations from -Os to make the default not that bad. this results in binaries that are only slightly larger but much faster:
http://git.musl-libc.org/cgit/musl/commit/?id=b90841e2583237a4132bbbd74752e0e9563660cd

rofl0r:
> I'll be running some further tests and inspection of the 16-bit code, especially with regards to eliminating the backwards conversion of left shifts to MUL instructions for higher speed on very slow 8088s.

cool. what were your findings ?

ghaerr:
> what were your findings ?

Super simple testing of left shift operations from this program:
```
/* optimize y * 80 */
int test2(int y)
{
    return (y << 6) + (y << 4);
}
```
a16-elf-gcc -melks-libc -mtune=i8086 -Os generates SHL's optimized to MUL (this is bad because there seems to be no way to turn this off and MUL is very very slow at 118 cycles on 8088):
```
00000000 <test2>:
   0:	55                   	push   %bp
   1:	89 e5                	mov    %sp,%bp
   3:	b8 50 00             	mov    $0x50,%ax
   6:	f7 66 04             	mulw   0x4(%bp)
   9:	5d                   	pop    %bp
   a:	c3                   	ret
```
a16-elf-gcc -melks-libc -mtune=i8086 -O2 or -O3 generates SHL's directly (this is what we want I think, using C code to suggest what to generate when writing for 8088):
```
00000000 <test2>:
   0:	89 e3                	mov    %sp,%bx
   2:	8b 47 02             	mov    0x2(%bx),%ax
   5:	d1 e0                	shl    %ax
   7:	d1 e0                	shl    %ax
   9:	d1 e0                	shl    %ax
   b:	d1 e0                	shl    %ax
   d:	89 c2                	mov    %ax,%dx
   f:	d1 e2                	shl    %dx
  11:	d1 e2                	shl    %dx
  13:	01 d0                	add    %dx,%ax
  15:	c3                   	ret
```
From the Intel docs, the cycles for just the two instruction MOV/MUL in the first example is 122 cycles, while the 8 instruction SHL/MOV/ADD is 23 cycles. So the changing from -Os to -O2 in this case is 5x quicker (!).

Admittedly for Pentium and beyond, -Os may be marginally quicker w/MUL at 11 cycles, but with faster CPUs none of this is a bottleneck anyways.

More testing coming to determine what else is changed when not running -Os. When compiled straight -Os, bin/nano-X was 68k. When compiled -O3, 77k, and with your suggested -O2+other -fstuff, 70k. So it looks like there's about a 12% difference in code size when running -Os vs -O3, but only a 3% increase when compiled -O2+. Pretty nice providing we can show what is running faster. The jury's still out on whether the gcc internal instruction timings used for opcode generation are different between 8086, 286, 386+ etc.

As described in my first post, I ran across a couple basic optimization settings that ended up producing code that crashed ELKS, but now can't get it to repeat. So until I get to the bottom of that, non -Os settings on the 16-bit ia16-elf-gcc are experimental.

PR #134, opened 2025-09-20
[driver] fix bit order of vga to vga blit on pc-98

tyama501:
Hello @ghaerr 

I needed to fix the shift of the vga to vga blit on pc-98
since the bit order of pc-98 vram is the reverse of the ibm pc.

This PR fixes the corruption of the first character on the left side of nxterm.

Before fix
 
Screenshot from 2025-09-20 16-41-30

After fix

Screenshot from 2025-09-20 17-14-08

Thank you!


ghaerr:
Thanks @tyama501. This is interesting, on the IBM PC there are sometimes some small vertical lines drawn outside the nxterm window when scrolling, depending on the position of the window. I had tried fixing this with various masking code, but will look again further now that you've identified and fixed this for PC-98.

PR #142, opened 2025-11-14
VGA X mode driver

toncho11:
A minimalistic VGA X driver for Nano X for ELKS.

The objective for now is to have something that compiles and works (even if barely). In another PR there will be optimizations for speed and memory consumption. Currently it compiles.

Here are some questions:
- is this the correct way to register the driver - there is SUBDRIVER and  PSUBDRIVER, which one to use?
- is the far pointer OK, I mean enough to make it work for now? Do I need to set psd->addr= ?
- how to set this driver as default one when building?
- other obvious problems?

### DIFF
--- a/src/drivers/vga_x.c
+++ b/src/drivers/vga_x.c
@@ -20,7 +20,9 @@ typedef unsigned short word;
 #define SCREEN_HEIGHT       200
 #define NUM_COLORS          256
 
-/* Far pointer to VGA memory */
+/* bytes per scanline seen by Mode X planar layout */
+#define BYTES_PER_LINE      80   /* 320 / 4 */
+
 static byte __far *VGA = (byte __far *)0xA0000000L;
 
 /* Set video mode using BIOS, used for open and close */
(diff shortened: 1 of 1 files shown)


toncho11:
VGA card on 8086/8088 is normal and I use it all the time. You just need to have [the list of compatible 16 bit ISA cards](https://forum.vcfed.org/index.php?threads/8bit-friendly-isa-vga-cards.26866/) that can run as 8 bit ISA VGA cards. Usually there is a jumper on the card for that. On ebay they often say if the card is 8 bit ISA compatible, which can make it a bit (but not too much) more expensive. Both my Amstrad and original IBM 8088 are on VGA cards.

Mode X is any unchained planar 256-color VGA mode, often but not always 320×240.

Resolution | VRAM needed
-- | --
320×200 | 64,000 bytes
320×240 | 76,800 bytes
360×240 | 86,400 bytes
320×400 | 128,000 bytes
400×300 | 120,000 bytes

Let's keep this PR and I will try to update it over time.

toncho11:
This needs to be totally rewritten. It will be a future project.

PR #151, opened 2025-12-11
Fix various exit problems with nxterm and nano-X

ghaerr:
Fixes a problem on the ELKS console after `nxterm` exits, the shell prompt is not displayed until ^C is typed. This required a bit more time spent in nxterm exiting for /bin/sh to change termios settings, as well as having nxterm call GrClose in all exit situations. The cursor is now always displayed on the last line of the console after nano-x exit.

Also turns on signal handling in ELKS `nano-x` which was causing it to crash when exiting nxterm in certain circumstances. Identified in https://github.com/toncho11/microwindows/issues/5.

@toncho11, this should fix both of your earlier reported problems of using `nxterm "edit /etc/mount.cfg && exit"`. Actually it's pretty cool to quickly be able to pull up a graphical terminal session and editor from the ELKS shell prompt. Things should go a lot smoother now.

Also previously `nxterm` changed its name to `-sh` in `ps` listings; this has been removed so it shows as `nxterm`. The shell running behind its pseudo terminal continues to run as `/bin/sh`, but with `/dev/ttyp0` as the terminal.

### DIFF
--- a/src/demos/nanox/nxterm.c
+++ b/src/demos/nanox/nxterm.c
@@ -1597,6 +1597,7 @@ static void sigpipe(int sig)
 	/* this one musn't close the window */
 	/*_write_utmp(pty, "", "", 0);  */
 	kill(-pid, SIGHUP);
+	GrClose();
 	_exit(sig);
 }
 
@@ -1604,6 +1605,7 @@ static void sigpipe(int sig)
 static void sigchld(int sig)
 {
 	/*  _write_utmp(pty, "", "", 0);  */
+	GrClose();
 	_exit(sig);
 }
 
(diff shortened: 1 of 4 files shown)


ghaerr:
> when you move the window you lose all the console contents. Maybe all characters need to be rewritten?

Yes. That's going to take a bit of work, since it doesn't currently have any redraw capabilities. I've made a note of it, but also have an idea for a more thorough terminal emulator written from scratch in my [GFX](https://github.com/ghaerr/gfx) project.

> once in nxterm (and nxterm is the only nx applciation) if you do "exit" then we have the old problem of not exiting nicely

Arrgh!! I thought I got things all fixed, I'll continue to play with fixing this. Agreed it would be nice to have a fully fast-start and fast-exit mechanism for using graphical apps in ELKS. We're almost there!

PR #154, opened 2025-12-13
Allow sending SIGTERM to all processes in nxstart pgroup to quit NX session

ghaerr:
This PR implements a way to gracefully exit from `nxstart` by adding a 'quit' menu entry. Also allows for other nxstart replacements a way to exit the entire Nano-X graphical system programmatically.

Discussed in https://github.com/ghaerr/elks/issues/2476, #152 and #153.

Now, nxstart calls `setsid()` at startup to create a new process group. `ps` will show each process's group, and all processes started from `nxstart`, including the nano-x server itself, will be in that process group. See the screenshot below, PGRP 13.

The method used is to send a SIGTERM (terminate) signal to all processes in the nxstart process group. This is done programmatically from the launcher using `kill(-getpid(), SIGTERM)`, or can be done from any shell using `kill -TERM -13` (for group 13).
ELKS nxstart

@toncho11, you should be able to use the following new code from nxstart.c in your launcher. See nxstart.c for full details:
```
/* Removes task entries after child exit and discards exit status */
static void reapchild(int sig)
{
    signal(SIGCHLD, reapchild);
    while(waitpid(-1, NULL, WNOHANG) > 0)
        continue;
}
    
static void exitwait(void)
{
    signal(SIGCHLD, SIG_IGN);
    GrClose();
    /*      
     * Wait for all children to exit before we do.
     * This prevents the shell we return to from reading
     * simultaneously with nxterm or another NX app and
     * causing bad behaviour.
     */
    while (waitpid(-1, NULL, 0) != -1)
        continue;
    exit(0);
}

/* handle SIGTERM by sending SIGTERM to all children, then wait for them to exit, then exit */
static void sigterm(int sig)
{
    signal(SIGTERM, SIG_IGN);
    kill(-getpid(), SIGTERM);
    exitwait();
}   
```
In main.c before GrOpen():
```
            
    signal(SIGCHLD, reapchild);
    signal(SIGTERM, sigterm);
    signal(SIGHUP, SIG_IGN);
    setsid(); 
```
Then to exit from the close box use:
```
    exitwait(); /* wait for all children to exit, then exit ourselves */
```
Or to quit forcefully:
```
    sigterm(SIGTERM); /* send SIGTERM to all children, then wait for them to exit, then exit */
```

### DIFF
--- a/src/demos/nanox/nxstart.c
+++ b/src/demos/nanox/nxstart.c
@@ -115,6 +115,7 @@ static void reapchild(int sig)
 		continue;
 }
 
+/* Wait for all children to exit, then exit ourselves */
 static void exitwait(void)
 {
 	signal(SIGCHLD, SIG_IGN);
@@ -130,6 +131,7 @@ static void exitwait(void)
 	exit(0);
 }
 
+/* Handle SIGTERM by sending SIGTERM to all children, then exit */
 static void sigterm(int sig)
 {
 	signal(SIGTERM, SIG_IGN);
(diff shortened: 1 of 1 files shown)


ghaerr:
> I do not understand if my program should have both 

The two functions are for completely different purposes. The first (1) above) purpose is required: when nxstart is closed via the close box, it must wait via `exitwait` for all other NX apps to be terminated by the user. When the user does that, then the Nano-X server exits to the shell. If exitwait is not called, then nxstart exits to the shell immediately, which then requires a ^C to see the shell prompt.

The `sigterm(SIGTERM)` functionality adds an option, currently shown as "Quit" in nxstart, which forces the entire set of NX applications include the nano-X server and nxstart itself, to terminate. The shell prompt is then displayed as usual afterwards.

toncho11:
Yes, I am writing my own replacement.
I am using sigterm(SIGTERM); for the exit item in my menu.
You made a pull request https://github.com/ghaerr/elks/pull/2527 I though it is related?
It all works now!!! Even CTRL + A exits successfully to the console.

toncho11:
Thank you!

PR #156, opened 2025-12-18
nxselect app

toncho11:
nxselect is a Nano X application that plays the role of a file selector. It is designed to be called by other applications.

When OK is clicked it returns the full path to a file.
If the app (the form) is closed then it returns "[]".

The nxselect plays a vital role in my main Nano X application that I intend to submit next. It is a required piece.


ghaerr:
Very nice, thank you @toncho11!

PR #158, opened 2025-12-19
nxdsktop

toncho11:
Ok so this is the big day. This is the application I have been working on. It is the culmination of all my efforts. 
No screenshot @ghaerr I want you to try it first! Do not even check the code.
Public announcement will be later.

### DIFF
--- a/src/demos/nanox/nxselect.c
+++ b/src/demos/nanox/nxselect.c
@@ -323,6 +323,18 @@ static void scroll_page_down(void)
         scroll_offset = max_offset;
 }
 
+static void set_title(GR_WINDOW_ID win, const char *title)
+{
+    GR_WM_PROPERTIES props;
+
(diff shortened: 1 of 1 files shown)


ghaerr:
@toncho11,

Wow!!! Very very very cool!!! You have really gone to town on this!! :) 

This is an amazing piece of work. I've already got some ideas and just tested the Restart and found nano-X can't be restarted. I think this may be due to the /tmp/nxsock file, I'll take a look at fixing that.

This brings a whole new era of usability to ELKS! 

The program seems to work quite well, I'll play around with the system more and post more comments.

What do you think about the idea of swapping the brown/orange background for the menu bar as well as the nxselect into a less-bright color, like blue, or blue-green? I can see the nano-X application default colors may need to be enhanced a bit. 

Thank you!!!

PR #178, opened 2026-05-05
Small updates fo nxedit.c

toncho11:
Updated comment section of nxedit.c.


ghaerr:
Thanks @toncho11!

toncho11:
Using nxedit takes 33 kb to open ./etc/profile while using the old method with nxterm + edit chained together uses 130 kb!  While still not fully tested this is a huge improvement. Arrows do not work on [copy.86 either. ](https://copy.sh/v86) either.

EDIT: Once we have a more stable and tested version of nxedit I will replace the "Edit file" start menu entry with nxedit.

PR #183, opened 2026-05-14
Fixed character corruption in nxedit

toncho11:
Fixed character corruption when using the arrow keys. 
Now only 1 character is repainted - the one where the old cursor was.
Several other modes I tested are still left. Let's keep them for now until the code is further tested.

For me that was a top priority fix - because I did not know if we can achieve a fast visual glitch free movement of the arrow keys and now we have it. 

Tested on copy.86 and 86box 286 emulation.

### DIFF
--- a/src/demos/nanox/nxedit.c
+++ b/src/demos/nanox/nxedit.c
@@ -74,11 +74,7 @@
 #define TITLE           "nxedit"
 #define COLS            80
 #define LINES           25
-/*
- * Do not define these as BLACK/LTGRAY directly: on some ELKS/Nano-X builds
- * those macros expand to large ARGB constants and cause -Woverflow warnings
- * when MWCOLORVAL is smaller. The real values are initialized at runtime.
- */
+
 #define FGCOLOR         fg_color
 #define BGCOLOR         bg_color
 
@@ -90,10 +86,6 @@
 
 /* ---------- compatibility fallbacks ---------- */
 
-/*
- * Some Nano-X trees use MWKEY_* and MWKMOD_* names. If your ELKS headers
- * differ, fix these aliases here rather than throughout the editor.
- */
 #ifndef MWKEY_ESCAPE
 #define MWKEY_ESCAPE    27
 #endif
@@ -192,13 +184,6 @@
  * It is useful when drawing the new vertical cursor damages pixels in the
  * character(s) to the right of the destination cursor.
  *
- * Best current guess:
- *   - do not repaint characters at the destination on normal cursor moves,
- *   - repair only the place the cursor left,
- *   - then draw the cursor at the new position.
- *
- * Enable this only if drawing the destination cursor causes permanent
- * damage to neighbouring characters after the cursor moves away.
  */
 #define CURSOR_DEST_REPAIR_ENABLE       0
 #define CURSOR_DEST_REPAIR_AFTER_DRAW   1
--- a/src/demos/nanox/nxedit.c
+++ b/src/demos/nanox/nxedit.c
@@ -74,8 +74,13 @@
 #define TITLE           "nxedit"
 #define COLS            80
 #define LINES           25
-#define FGCOLOR         BLACK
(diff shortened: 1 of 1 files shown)


ghaerr:
Hi @toncho11, glad you got the drawing fixed after using any arrow key, I hadn't noticed that.

I had just made a commit to fix improper use of 'int' for specifying colors rather than MWCOLORVAL, and now your PR overwrites that, along with more code that was totally incorrect for "getting rid of compiler warnings". I see that a 2nd commit got rid of most of that, but there's still some pretty unneeded and incorrect code, which I'll comment on separately. I had pushed the fix since the previous code produced tons of errors on the output stream, which then makes other Nano-X code hard to maintain, since the developer then doesn't know whether their code or old code has errors, when they see bold font scrolling quickly on the screen.

It appears your AI completely misunderstands the fact that colors are specified as 32-bit values in Nano-X, as it then writes code to pass them to functions using 'int' which is 16-bit. Likely it doesn't know its writing for ia16-elf-gcc, but I have no idea. I'm not really much into debugging AI code, TBH. When int and MWCOLORVAL are confused, this causes compiler warnings that I don't like to have in Nano-X, since I spent so much time removing them in the past.

This business of using AI to write code that is not tested is troubling me, I'm considering going the way Linux decided and requiring "Assisted by: model, version" in the source so that others know that it is entirely possible that nobody actually understands how the program works. I also think we should require all submissions to at least have been compiled by the submitter and no warnings are present. These kinds of things are becoming more apparent to me after looking at more code over at ELKS. While I don't mind AI code, I can see it's a good way to introduce unknown bugs since very little testing or real understanding is occurring. What do you think about this?

I'll indicate the changes required for this to be committed, and then comment on items that should be un-rewritten and returned to the way they were better written the first IMO.

Thank you!

src/demos/nanox/nxedit.c
```diff
@@ -844,7 +911,7 @@ static int line_baseline_y(unsigned short line_no)
     return TOP_MARGIN + ((int)line_no - first_line) * LINE_HEIGHT;
 }
 
-static void set_fg(MWCOLORVAL color)
+static void set_fg(int color)
```
ghaerr, line 914: Change back to MWCOLORVAL, and also on line 1021.

src/demos/nanox/nxedit.c
```diff
@@ -73,8 +74,9 @@
 #define TITLE           "nxedit"
 #define COLS            80
 #define LINES           25
-#define FGCOLOR         BLACK
-#define BGCOLOR         LTGRAY
+
+#define FGCOLOR         fg_color
+#define BGCOLOR         bg_color
```
ghaerr, line 79: These two lines 78-79 should be deleted and 76-77 restored, along with deleting the function `make_color` on lines 251-253 and 256-264. It's all extra garbage, when using BLACK and LTGRAY are much more straightforward, which demand proper typing for set_fg using MWCOLORVAL.

Also delete calls to `make_color` on lines 1606-7.

Those changes essentially undo the needless move to fg_color and fg_color variables and the needless function make_color which is passed numbers rather than just spelling out the color desired. When a user or system color change is desired, believe me it will be much easier this way.

toncho11:
My job was to make sure the arrows work correctly. I designed several strategies. I compiled and tested them. I found a solution. I do not have time for what you request. It  you want you can fix these after my pull request or not.

Check if there is an arrow keys roblem with the old code, there were definitely glitches and I was confused how you did not noticed.

ghaerr:
> It you want you can fix these after my pull request or not.

Allright, I'll make the fixes. I haven't had much time myself lately, but do appreciate your enhancement to nxedit. There was definitely a problem with arrow keys in the old code - I had already committed that PR before I was able to notice.

toncho11:
Thank you.

ghaerr:
@toncho11: The suggested cleanups were made in afb26b3a704a2c689869aff51eac097e9eccfafb. Please do a git pull before more modifications, thanks.

PR #188, opened 2026-05-25
nx edit imrovement

toncho11:
- only one policy for handling the cursor movement left. This reduced the code.
- for each nxedit instance a different swap file is created using the nxedit's PID. It also handles if the file name already exists. It tries to create a nxedit's PID + i.
- the swap file is deleted on exit


ghaerr:
Thanks @toncho11!

PR #193, opened 2026-07-22
Add NuttX RTOS screen, keyboard and mouse/touchscreen drivers

Acfboy:
Introduce scr_nuttx.c using the NuttX framebuffer API, keyboard drivers for event-mode (kbd_nuttx_event.c) and raw byte-stream (kbd_nuttx_raw.c), and pointing drivers for mouse (mou_nuttx_mouse.c) and touchscreen (mou_nuttx_ts.c).  Each driver accepts a device path override.

Add mwconfig.nuttx with NuttX-appropriate defaults, included via the MWCONFIG_FILE mechanism.  Add NuttX ARCH/KEYBOARD/MOUSE entries to Objects.rules so the correct drivers are linked when building for NuttX.

### DIFF
--- a/src/drivers/Objects.rules
+++ b/src/drivers/Objects.rules
@@ -132,6 +132,23 @@ MW_CORE_OBJS += $(MW_DIR_OBJ)/drivers/mou_ajaguar.o
 MW_CORE_OBJS += $(MW_DIR_OBJ)/drivers/kbd_ajaguar.o
 endif
 
+# NuttX RTOS
+ifeq ($(ARCH), NUTTX)
+MW_CORE_OBJS += $(MW_DIR_OBJ)/drivers/scr_nuttx.o
+ifeq ($(KEYBOARD), NUTTXKBD)
+MW_CORE_OBJS += $(MW_DIR_OBJ)/drivers/kbd_nuttx_event.o
+endif
+ifeq ($(KEYBOARD), NUTTXRAWKBD)
+MW_CORE_OBJS += $(MW_DIR_OBJ)/drivers/kbd_nuttx_raw.o
+endif
+ifeq ($(MOUSE), NUTTXMOUSE)
+MW_CORE_OBJS += $(MW_DIR_OBJ)/drivers/mou_nuttx_mouse.o
+endif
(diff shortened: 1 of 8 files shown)


ghaerr:
> Could you give me some advice?

Looks great and very nicely done! Would you like this committed now?

Acfboy:
Thank you for your quick reply. If @ppsia also sees no issues, I think merging it should be a good idea, and I will soon open a formal PR in nuttx-apps.

ppisa:
@Acfboy and @ghaerr thanks for the work. I have only gone through code quickly I return to it again. But it looks well.

It would be great if other NuttX developers have time to look at it, @acassis, @michallenc etc.

I hope that the keycodes in NuttX are resolved as well. But with my proposal to shift codes for special keys above ASCII range, the provided code should work.

ghaerr:
To keep things moving forward, I can commit this now, with possible changes after more review by others? This would allow for full-system building including pulling from microwindows repo. I don't mind receiving updates for NuttX-specific changes at any time.

ppisa:
Thanks, I would wait for one day at least if there is some feedback from somebody other from NuttX community.
If not and it is not problem for you that there would be more fixup patches then it would be great to have code in official Microwindows repository.

As for accompanying NuttX applications integration commit

https://github.com/Acfboy/nuttx-apps/commit/fb8d2389631af8d6e01330ef2b6299f17a1f0b0f

I would add options resulting in `NOMOUSE` and `NOKBD` in `Kconfig`

```
+choice
+	prompt "Keyboard driver"
+	default MICROWINDOWS_KBD_EVENT
...
+
+config MICROWINDOWS_KBD_EVENT
+	bool "Event-mode keyboard driver"
...
+
+config MICROWINDOWS_KBD_RAW
+	bool "Raw-mode keyboard driver"
```

It is quite possible that there is hardware which does not use mouse or KBD or even such which is used to display only some information. There could be even some "Custom" option which does not request driver from Microwindows and application or BSP provides its own local solution.

ppisa:
As there are no more comments, I thing that it worth to merge it to mainline if you agree and updates and fixes can be prepares as the development goes. But I think this is reasonable base.

PR #195, opened 2026-07-29
[driver] NuttX keyboard fixed to recognize special keys from event type

ppisa:
This is initial fix to support both special keys and alphanumeric keys
in the Microwindows demo edit widgets.

This does not solve still application of the shift and capslock through modifiers.

Related NuttX problems with omitted delivery of special keys as KBD_SPECPRESS
and KBD_SPECREL for multiple keyboards is tracked at

https://github.com/apache/nuttx/issues/19527

### DIFF
--- a/src/drivers/kbd_nuttx_event.c
+++ b/src/drivers/kbd_nuttx_event.c
@@ -79,7 +79,9 @@ static void nuttxkbd_GetModifierInfo(MWKEYMOD *mods, MWKEYMOD *curmods)
 static int nuttxkbd_Read(MWKEY *kbuf, MWKEYMOD *mods, MWSCANCODE *scancode)
 {
   struct keyboard_event_s event;
+  MWKEY key;
   int n;
+  int press;
 
   n = read(kbd_fd, &event, sizeof(event));
   if (n < (int)sizeof(event))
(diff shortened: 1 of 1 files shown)


ppisa:
@Acfboy please, look at my changes and test them.

By the way, it seems that
```
HAVE_FNT_SUPPORT = Y
```
has to be provided at least by `nuttx-apps/graphics/microwindows/Makefile` for correct build and linking. It probably can be omitted if it is removed from `microwindows/src/include/mwconfig.nuttx`. But we probably want to have it there for now.

probably the most flexible option would be to generate `mwconfig` fragment included through `MWCONFIG_FILE` at build time from NuttX Kconfig. But it would be too much complications for now.

ghaerr:
Thank you @ppisa. Please let me know when you want this committed.

Acfboy:
> @Acfboy please, look at my changes and test them.

It looks good. I have tested them locally with https://github.com/apache/nuttx/pull/19577, and I can type space and backspace in `sim:mw` now.

ppisa:
Thanks to @Acfboy for testing. @ghaerr please push it to master. It is possible that there come more changes as the NuttX side is enhanced but I think that even this patch helps to move forward.

PR #196, opened 2026-08-10
Add NX_PER_CLIENT_DATA support for NuttX and eCos

Acfboy:
In a flat build the client library and the server share a single address space and several client tasks can connect to the server at the same time. All per-client statics of the client library are therefore moved into a structure whose pointer is kept in thread-local storage, and controlled by a standard MULTITHREAD_SERVER define now.


ghaerr:
Hello @Acfboy,

In general this looks very good, thank you. I have a few small comments I'll place next to the source lines.

I have one big concern which is duplicating the entire function nxPaintNCArea in a separate file nxpaintnc.c. When I suggested separating that function from nxdraw.c, I was hoping to see the function removed from nxdraw.c also, and not having the entire nxpaintnc.c file \#if MULTITHREAD_SERVER. Maintenance is much harder when a function is duplicated in its entirety in separate source files. Can you make it work by optionally including serv.h in nxpaintnc.c only for the MULTITHREAD_SERVER case, thus having only a single copy of the function source? If so, please go ahead and do that.

Thank you!

src/include/mwconfig.h
```diff
@@ -134,6 +134,10 @@
 #define NONETWORK		0		/* =1 to link Nano-X apps with server for standalone*/
 #endif
 
+#ifndef MULTITHREAD_SERVER
+#define MULTITHREAD_SERVER	0	/* =1 for per-client data in a single-process multi-threaded Nano-X build (eCos or POSIX thread-local storage)*/
```
ghaerr: Please shorten the comment to 90 columns.

src/nanox/client.c
```diff
@@ -27,6 +27,7 @@
 #include <errno.h>
 #include <time.h>
 #include <assert.h>
+#include <pthread.h>
```
ghaerr: Lets move this pthread.h include to the MULTITHREAD_SERVER case below so that systems without it still compile.

src/nanox/client.c
```diff
@@ -100,14 +94,38 @@ static EVENT_LIST *	evlist;
...
+
+#if MULTITHREAD_SERVER && !__ECOS
+pthread_once_t g_nanox_data_key_once = PTHREAD_ONCE_INIT;
+pthread_key_t  g_nanox_data_key;
+
+void make_nanox_data_key(void)
+{
+  pthread_key_create(&g_nanox_data_key, free);
+}
+#endif
```
ghaerr: Can you move lines 119-126 to an \#else clause above on 107? This will be less messy that way I think.

src/nanox/Makefile
```diff
@@ -47,6 +47,11 @@ OBJS := \
 	$(MW_DIR_OBJ)/nanox/nxtransform.o
 OBJS += $(NANOWMOBJS)
 
+ifeq ($(NANOWM), Y)
+# Server-side non-client area painting for the built-in window manager
+NANOXFILES += $(MW_DIR_OBJ)/nanox/nxpaintnc.o
+endif
```
ghaerr: Add nxpaintnc.o to line 45 above also once the contained function is removed from nxdraw.o.

I guess this separate line is required here, since we're not building a client library but instead the server, right?

src/nanox/nxpaintnc.c
```diff
@@ -0,0 +1,128 @@
...
+ * This file provides the server-side counterpart of nxPaintNCArea() from
+ * nxdraw.c.  It includes serv.h, so that nxPaintNCArea and the GrXXX
+ * calls are renamed to the server-side SVR_nxPaintNCArea / SVR_GrXXX
+ * symbols, just like the other server files.  nxdraw.c keeps its
+ * client-side nxPaintNCArea() for client applications and for builds
+ * without MULTITHREAD_SERVER.
+ */
+#if MULTITHREAD_SERVER
+#include "serv.h"
+#include "nanowm.h"
```
ghaerr, line 18: Perhaps just add an \#endif here as discussed previously, and change the comment above. Hopefully that will work.

src/Arch.rules
```diff
@@ -221,7 +221,7 @@ ifeq ($(ARCH), ECOS)
 TOOLSPREFIX = $(ARMTOOLSPREFIX)
 ECOS_PREFIX = /work/iPAQ/ecos/install
 INCLUDEDIRS += -I$(ECOS_PREFIX)/include
-DEFINES += -D__ECOS=1 -DUNIX=1 -DHAVE_SIGNAL=0 -DHAVE_MMAP=0 -DHAVE_SELECT=0
+DEFINES += -D__ECOS=1 -DUNIX=1 -DHAVE_SIGNAL=0 -DHAVE_MMAP=0 -DHAVE_SELECT=0 -DMULTITHREAD_SERVER=1
```
ghaerr: Lets break this into two DEFINES += lines so not seeing auto-linewrap.

Acfboy:
Thanks, fixed.

ghaerr:
This looks great, thanks for the changes. Do you want a review from the people on your side? Let me know when you'd like to commit, thank you!

Acfboy:
Perhaps we should wait until I submit the PRs in nuttx and nuttx-apps, and then let @ppsia take a look as well.

src/include/mwconfig.nuttx
```diff
@@ -2,7 +2,10 @@
  * Included by mwconfig.h via -DMWCONFIG_FILE='"mwconfig.nuttx"'
  */
 
-#define NANOWM              1
+#ifndef NANOWM
```
ppisa, line 7: I would suggest to minimize dependency on external defines there as I write in https://github.com/apache/nuttx/issues/18566#issuecomment-5266692452 . If there is only single externally configurable option required then (may it be use) have two separated config files , one for case with Nano-WM `mwconfig.nuttx-nanowm` and another without it ``mwconfig.nuttx`` or to leave this file there as a template/example and generate another one during config. But I agree that neither is ideal.

src/include/mwconfig.nuttx
```diff
@@ -2,7 +2,10 @@
  * Included by mwconfig.h via -DMWCONFIG_FILE='"mwconfig.nuttx"'
  */
```
ppisa, line 4: I would consider to add there
```
#undef  NUTTX
#define NUTTX               1
```
It should propagate to exported headers this way even when the number of exported defines is kept on minimum, i.e. only
```
CFLAGS += -DMWCONFIG_FILE='"mwconfig.nuttx"'
```

ppisa (commented):
As for `MULTITHREAD_SERVER`, there are more changes. As I understand that should be safe because it should not have potential to break anything else than ECOS which is not actively maintained anyway. So no damage for most used targets.

I have still some taste that `MULTITHREAD_SERVER` is not the right name. But I am not sure if I am right. I have feeling that it should be something like `NANOX_PER_THREAD_CONNECTION` or something similar because I think that this is main purpose. But I can be wrong and the @ghaerr opinion and consideration about disruptiveness of the changes are the most important.

Acfboy:
Sorry, @ghaerr. I was offline for a while due to some personal matters. I’ve added the NUTTX flag and  it no longer polls when NUTTX is enabled. How does it look now?

src/mwin/winmain.c
```diff
@@ -25,6 +25,7 @@
 #include "windlg.h"
 #include "device.h"
 #include "osdep.h"
+#include "mwconfig.h"
```
ghaerr: device.h already includes mwconfig.h, can you remove this extra include?

ghaerr:
@Acfboy, looks good with the one remaining change removing mwconfig.h in winmain.c.

Let me know when you'd like to commit.

Acfboy:
@ghaerr fixed. I think it can be merged now, and I can update the commit hash in Makefile on the nuttx-apps side.

PR #198, opened 2026-08-22
Fix some lint issues in nuttx

Acfboy:
Hi, @ghaerr . Could we fix some ci `-Werror` issue mentioned in https://github.com/apache/nuttx/pull/19912 here?

- Guard the memplan4 framebuffer subdriver declaration with #if ELKS so it is not an unused variable on non-ELKS builds.
- Convert the space-indented statements in GdRasterOp to tabs to fix the misleading-indentation warning.
- Declare FARPROC, NEARPROC and PROC with (void) parameters to fix the strict-prototypes warning.

### DIFF
--- a/src/drivers/fb.c
+++ b/src/drivers/fb.c
@@ -105,7 +105,9 @@ select_fb_subdriver(PSD psd)
 	extern PSUBDRIVER fblinear24[4];
 	extern PSUBDRIVER fblinear32bgra[4];
 	extern PSUBDRIVER fblinear32rgba[4];
+#if ELKS
 	extern PSUBDRIVER memplan4[4];
+#endif
 
 #if ELKS
 	pdriver = memplan4;
--- a/src/drivers/rasterops.c
+++ b/src/drivers/rasterops.c
@@ -307,35 +307,35 @@ GdRasterOp(PMWIMAGEHDR pixd, MWCOORD dx, MWCOORD dy, MWCOORD dw, MWCOORD dh, int
 	/* translate op to pix library format*/
 	op = raster_ops[op];
 
-    if (op == PIX_DST)   /* no-op */
-        return;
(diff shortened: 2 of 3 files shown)


ghaerr:
> Could we fix some ci -Werror issue mentioned in https://github.com/apache/nuttx/pull/19912 here?

Sure. I can see there's more warnings there than I would have thought. I may clean up some of the others seen when I have time.

> Declare FARPROC, NEARPROC and PROC with (void) parameters to fix the strict-prototypes warning.

I don't see any actual uses of FARPROC or NEARPROC, but not fully sure about PROC. In some cases, adding a (void) parameter list could cause problems should their current empty () be required to allow a variety of arguments, but lets go ahead and change this now, as if this issue comes up it can always be changed back.

Thanks!

ppisa:
As I mention in https://github.com/ghaerr/microwindows/pull/194#issuecomment-5422485762 , the relatively mild intrusive solution to undefined macros warning is use of the next pragma at start of header files and C files where it is necessary
```
#pragma GCC diagnostic ignored "-Wundef"
```
and I expect that large part would be solved by adding it at start of `mwconfig.h` or it can be put even to `mwconfig.nuttx` but I think that its use for all targets would be OK and advantage.

ghaerr:
@ppisa,

> #pragma GCC diagnostic ignored "-Wundef"

If this works for NuttX, that's great, lets add it.

> it can be put even to mwconfig.nuttx 

Since the problem only occurs with the NuttX configuration requiring -Wundef, I would prefer solving the problem using only mwconfig.nuttx, as it seems that the real issue here is that -Wundef is incompatible with the way that Microwindows uses symbols for \#if directives. This leaves the rest of the compilations as-is, where no warnings are currently produced.

Please submit a PR if this sounds good.

PR #201, opened 2026-09-09
scr_nuttx.c: call FBIO_UPDATE ioctl when required by the display

Vajnar:
On NuttX some displays require explicit update notification on content change. This commit adds respective `ioctl()` invocation. This commit was made with help from @ppisa .

EDIT: Add also commit with mapping between NuttX and Microwindows pixel types.


ppisa:
Thanks and please send even mapping of screen pixel format to MW format to add RGB565 and more in addition to ARGB8888. The code has been checked on EPS32C6 and and it would be ideal to test it on your STM32 LCD equipped board as well.

Vajnar:
Hi @ppisa , I tested the pixel types mapping on my STM32 and the example application is displayed correctly. I therefore add the pixel type mapping to this PR.

ppisa:
Thanks, I have rechecked on qemu-intel64 and everything works correctly.

@ghaerr please, merge changes to mainline. They are only local enhancements to the NuttX screen driver.

